PR #14, opened 2018-02-24
decompile main

ProjectRevoTPP:


src/main.c
```diff
@@ -0,0 +1,105 @@
+#include "global.h"
+
+extern int sub_8001C5C();
```
camthesaxman: Please use functions.h for externs. Just uncomment the function and give it the appropriate declaration. We want to avoid having all these externs in the .c files.

src/main.c
```diff
@@ -0,0 +1,105 @@
...
+extern void SetInterrupt_LCDVBlank(void *func);
+extern void ResetKeyStatus(void *keyPtr); // TODO: Uses a struct defining the key info. Define this struct.
+extern void UpdateKeyStatus(void *keyPtr); // same as above
+extern void StoreRoutinesToIRAM();
+extern void sub_80A2C3C();
+extern void Proc_Initialize();
```
camthesaxman: This is in proc.h

src/main.c
```diff
@@ -0,0 +1,105 @@
...
+extern void GeneralVBlankHandler(void);
+extern void sub_80BC81C(void);
+extern void SetSomeByte(int);
+extern void Font_InitForUIDefault(void);
+extern void NewGameControl(void);
+extern void ExecMainUpdate();
+extern void sub_8001C78();
+extern void PrintDebugStringToBG(u16 *bg, u8 *asciiStr);
+
+extern void **gUnknown_0858791C; // contains a pointer to key area (TODO: Define Key Info Struct)
```
camthesaxman: Same with this. use variables.h instead of putting the extern here.

PR #38, opened 2018-03-01
generate compressed string data

camthesaxman:


ProjectRevoTPP:
Test

ProjectRevoTPP:
test2

laqieer:
# Idea on how to match Huffman table
First, implement a Huffman encoder and generate the Huffman tree.
- If it matches, we are done.
- If it doesn't match, there are two possible reasons:
  1. the generated Huffman tree is not unique
  2. the original tree is not generated from the final text

Then calculate the WPL of generated Huffman tree and original tree.
- If they are not equal, it proves to be reason ii, then we can attach extra unused text to adjust the frenquency to match it.
- If they are equal, it may be reason i, then we try to rotate nodes at the same level or tiebreaks handle logic to match it.

laqieer:
# Idea on how to match Huffman table
First, implement a Huffman encoder and generate the Huffman tree.
- If it matches, we are done.
- If it doesn't match, there are two possible reasons:
  1. the generated Huffman tree is not unique
  2. the original tree is not generated from the final text

Then calculate the WPL of generated Huffman tree and original tree.
- If they are not equal, it proves to be reason ii, then we can attach extra unused text to adjust the frenquency to match it.
- If they are equal, it may be reason i, then we try to rotate nodes at the same level or tiebreaks handle logic to match it.

laqieer:
I think the original tree is the optimal tree from the final text in game, because it has the totally same WPL (4193029) as our generated tree.

laqieer:
# Guess
The main steps of Huffman encoding: read texts -> count frequency -> merge node.
- If the game developer split texts into different files with their own filenames or grouped into different folders, their process order will be not sure.
- The frequency count is usually implemented using dictionary. It is unordered, so its visit order is not sure either.
- The node merge is usually implemented using priority queue. It is based on heap sort, which is also unordered, so the order of nodes with same value is not sure either.

PR #50, opened 2018-04-02
Split code into more files

StanHash:
- end of `bmdebug.s`
- `playerphase.s` Player Phase proc code and related functions
- `koido.s` KOIDO and KOIDOAMM procs (rescue/drop/give/take map unit animations)
- `bmxfade.s` Tile Fading (for tile changes and such)
- `bmcamadjust.s` Misc Camera movement functions
- `convoymenu.s` "Send item to convoy" menus and related functions
- `menuitempanel.s` Item info panel in item selection menus
- `unitswapfx.s` Unit Swapping Animation/Effect (When you swap unit places in the prep screen)
- `phasechangefx.s` The nice phase changing thing and related procs/functions
- `trapfx.s` Various Graphical Effects related to activating traps
- `notifybox.s` Some kind of inferior popup (used for trap removing among other things)

Note: I'm assuming `bm` stands for "battle map".

There were some weird out of place functions here and there so I put those in their own (`code_xxxxxxx.s`) files for now.

Hopefully I didn't split *too* much. I tried to make sure the (ro)data had corresponding boundaries (no overlaps) as best as I could. If it happens to not match later I guess we'll have to merge some files together but for now I think it's okay.

### DIFF
--- a/asm/bmcamadjust.s
+++ b/asm/bmcamadjust.s
@@ -1,0 +1,188 @@
+	.INCLUDE "macro.inc"
+
+	.SYNTAX UNIFIED
+
+	THUMB_FUNC_START GetPlayerStartCursorPosition
+GetPlayerStartCursorPosition: @ 0x0801DE30
+	push {r4, r5, r6, lr}
+	adds r5, r0, #0
+	adds r6, r1, #0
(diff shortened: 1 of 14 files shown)


ProjectRevoTPP:
fix conflict

PR #58, opened 2018-09-11
decompile mu.s into mu.c

StanHash:
~~*it's alive!*~~

One functions is non-matched (`MU_IsAnyActive`, which is somehow one of the smaller functions of the file) because it hates me apparently.

All data that seems to correspond to this file has also been added in; Including all the movement sound definitions (that may need to be moved into some separate (maybe generated?) header eventually?).


src/mu.c
```diff
@@ -0,0 +1,1951 @@
...
+    return Proc_Find(gProcScr_MoveUnit) ? TRUE : FALSE;
+}
+
+#if NONMATCHING
+
+int MU_IsAnyActive(void) {
+    int i;
+
+    for (i = 0; i < MU_MAX_COUNT; ++i)
+        if ((sMUConfigArray[i].muIndex) && (sMUConfigArray[i].pMUProc->stateId != MU_STATE_NONACTIVE))
```
ProjectRevoTPP: Have you tried breaking up the if statements to multiple nested ones?

src/mu.c
```diff
@@ -0,0 +1,1951 @@
...
+    CopyToPaletteBuffer(palette, (0x20 * (0x10 + proc->pMUConfig->paletteIndex)), 0x20);
+}
+
+#if NONMATCHING
+
+void MU_SetPaletteId(struct MUProc* proc, int paletteId) {
+    proc->pMUConfig->paletteIndex = paletteId;
+
+    proc->pAPHandle->tileBase =
+        ((proc->pMUConfig->paletteIndex & 0xF) << 12) + proc->pMUConfig->objTileIndex + proc->objPriorityBits;
```
ProjectRevoTPP: bitfield?

PR #66, opened 2019-01-09
decompile bmunit

StanHash:
No nonmatches. Also added class & character constants header files.


include/bmitem.h
```diff
@@ -1,7 +1,51 @@
...
+    u8 encodedRange; //19
+
+    u16 costPerUse; //1a
+    u8  weaponRank; //1c
+    u8  iconId; //1d
+    u8  useEffectId; //1e
+    u8  weaponEffectId; //1f
+    u8  weaponExp; //20
+
+    u8  _u21[3]; //21
```
ProjectRevoTPP: could be struct padding

src/bmunit.c
```diff
@@ -1,3 +1,1612 @@
...
+        u16 monsterClass = GenerateMonsterClass(uDef->classIndex);
+
+        buf = *uDef;
+
+        buf.autolevel = TRUE;
+        buf.classIndex = monsterClass;
+        buf.level = GenerateMonsterLevel(uDef->level);
+
+        {
+            u32 packedItems = GenerateMonsterItems(monsterClass);
```
ProjectRevoTPP: im assuming the var declared down here is needed to match

PR #94, opened 2019-05-05
fix and clean up ROM padding and split out more end ROM objects

ProjectRevoTPP:


Makefile
```diff
@@ -10,14 +10,10 @@ CC1      := tools/agbcc/bin/agbcc$(EXE)
 CC1_OLD  := tools/agbcc/bin/old_agbcc$(EXE)
 #include $(DEVKITARM)
 PREFIX = arm-none-eabi-
-export CPP := cpp
```
ProjectRevoTPP: I thought I undid this change

PR #102, opened 2019-08-15
Work around compress linker EABI mismatch errors

StanHash:
I was getting a bunch of those errors:

    [...]/arm-none-eabi-ld: error: source object data/banim/[...].o has EABI version 5, but target data/banim/data_banim.o.bak.o has EABI version 0

Which made banims fail to build entirely, which in turn made everything fail to build entirely :/

~~I managed to fix it (for me at least) by adding a `-B armv4t` to the objcopy invocation where there wasn't one in the compressing linker.~~ See comment below.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -80,7 +80,8 @@ include graphics_file_rules.mk
 %.8bpp: %.png  ; $(GBAGFX) $< $@
 %.gbapal: %.pal
 ifneq ($(OS),Windows_NT)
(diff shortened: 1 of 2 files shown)


StanHash:
actually I lied that didn't work, what worked was adding `--no-warn-mismatch` to the ld invoke which is more of a workaround than a fix :/ :/

PR #106, opened 2019-12-11
decompile statscreen

StanHash:
All good. A lot of data was also extracted.


src/statscreen.c
```diff
@@ -0,0 +1,3064 @@
...
+        TEXT_COLOR_BLUE, stats->winAmt);
+
+    // Display Loss Amt
+    sub_8004B88(gBmFrameTmap0 + TILEMAP_INDEX(11 + sub_80AEBEC(stats->lossAmt), 14),
+        TEXT_COLOR_BLUE, stats->lossAmt);
+}
+
+#ifdef NONMATCHING
+
+void DrawStatScreenBar(int num, int x, int y, int base, int total, int max)
```
ProjectRevoTPP: Whats nonmatching about this?

StanHash: Looking at it closely again and I got it to match.
It was, of course, something dumb (the `total` argument needed to be loaded into a register which I initially thought was something to do with the diff variable being weird but actually that wasn't it).

StanHash:
Now it is done

PR #131, opened 2021-12-05
Decompile asm/bmphase.s

NoahNelson:
Also fix a spelling mistake and rename a function


src/bmphase.c
```diff
@@ -0,0 +1,84 @@
...
+        ret = 0;
+    };
+    return ret;
+}
+
+int GetCurrentPhase(void) {
+    return gRAMChapterData.chapterPhaseIndex & ENEMY_PHASE_INDEX;
+}
+
+int GetActiveNonEnemyFaction(void) {
```
NoahNelson, line 82: oh, when I think about it a bit more this name doesn't really make sense.

PR #142, opened 2022-02-17
Rename BmMapUnk to BmMapOther.

laqieer:
Thanks @MokhaLeee for the information.


StanHash:
that map is used for much more than just the magic seal
it is used, for instance, as the ai "danger" map.

I call it simply `g[Bm]MapOther` in fe6

laqieer:
Thanks for the information.

PR #143, opened 2022-02-22
Fix: LF line endings aren't supported.

laqieer:
Always convert line endings to CRLF on checkout for `.pal` files, because they must keep CRLF endings, even on OSX or Linux, otherwise gbagfx will complain that `LF line endings aren't supported.`.
https://github.com/FireEmblemUniverse/fireemblem8u/blob/c90d89062ff25c7991efa5385f3959a2d339159c/tools/gbagfx/jasc_pal.c#L49

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,6 +1,9 @@
 # Auto detect text files and perform LF normalization
 * text=auto
 
+# Declare files that will always have CRLF line endings on checkout.
+*.pal text eol=crlf
+
 # Custom for Visual Studio
 *.cs     diff=csharp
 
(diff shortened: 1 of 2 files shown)


laqieer:
2 solutions to fix this problem:
- use `git` to convert the line endings
https://github.com/pret/pokeemerald/blob/eca5233abe83a074ce97f9e47b30eb0c0f430a9e/.gitattributes#L1
- fix `gbagfx` to accept LF line endings
https://github.com/laqieer/fe7-jp-stunning-tribble/commit/b58dcc7c44bae37d351ec57c86f603bf695bf9e3
I choose the first one to be consistent with pret.

PR #146, opened 2022-03-07
Decompile ui_targetselection into uiselecttarget

Eebit:
Tried to match the convention from #103 and #104.

Still trying to get the hang of this, so please feel free to let me know if there are any glaring mistakes here. Would appreciate help and guidance around the non-matches as well!


src/uiselecttarget.c
```diff
@@ -0,0 +1,483 @@
...
+    }
+
+    gKeyStatusPtr->newKeys = 0;
+
+    return proc;
+}
+
+#if NONMATCHING
+
+ProcPtr NewTargetSelection_Specialized(const struct SelectInfo* selectInfo, int(*onSelect)(ProcPtr, struct SelectTarget*)) {
```
ProjectRevoTPP, line 224: How is this one not matching? Do you have a scratch or CE link for this one?

Eebit: Aside from the fact that I forgot to change the return type of the function pointer here (fixed in 349ddf5)...  :) 

I could not get it to use the extra register when setting `onAPress`:

```diff
	push {r4, lr}
	adds r4, r1, #0
-	bl NewTargetSelection
-	adds r1, r0, #0
-	str r4, [r1, #0x38]
+ 	bl NewTargetSelection
+	str r4, [r0, #0x38]
	pop {r4}
	pop {r1}
	bx r1
```

Sorry, but I'm not familiar with what you mean by a scratch or CE link! Could you point me to any materials I could use to learn?

Eebit: Ah - joined the Discord server and figured it out in the end. Here's the link: https://cexplore.karathan.at/z/ofW2pR

ProjectRevoTPP:
Update this

PR #151, opened 2022-05-10
Decompile chapterintrofx

Eebit:
Borrowed the naming conventions and some logic from FE6 decomp (thanks Stan!!) and also some supporting notes from @Huichelaar's [FE8U_ChapterIntro](https://github.com/Huichelaar/FE8U_ChapterIntro) to get my bearings.

A couple non-matches here...

`ChapterIntro_InitMapDisplay` (formerly `sub_8020B30`) failed to match even though I am pretty sure I followed [the equivalent function from FE6](https://github.com/StanHash/fe6/blob/0b25087de406c04502d0e50d5ac184d54bbbae5b/src/bmfx.c#L1986-L2018). There is an extra use of `r2` that I couldn't seem to shake.

`ChapterIntro_TickTimerMaybe` (formerly `sub_8021090`) is using the wrong register.

I'll add some Compiler Explorer links here later.


src/chapterintrofx.c
```diff
@@ -0,0 +1,1344 @@
...
+    DisableMapPaletteAnimations();
+
+    UnpackChapterMapGraphics(gRAMChapterData.chapterIndex);
+
+    SetupMapSpritesPalettes();
+    LoadObjUIGfx();
+
+    // TODO - Nonmatching is caused by an extra use of r2 around here for the results of this operation
+
+    var = GetROMChapterStruct(gRAMChapterData.chapterIndex)->_unk10[2];
```
ProjectRevoTPP, line 898: did you try encapsulating these areas in do{}while(0)s and locally initializing var in each scope?

src/chapterintrofx.c
```diff
@@ -0,0 +1,1344 @@
...
+    return;
+}
+
+#if NONMATCHING
+
+void ChapterIntro_TickTimerMaybe(struct ChapterIntroFXProc* proc) {
+    if (proc->unk_52 != 0) {
+        Proc_Break(proc);
+        return;
+    }
```
ProjectRevoTPP, line 1251: Is the return here required? Maybe move the decrement into the if and make it an else if instead of using a return?

PR #156, opened 2022-06-22
Decompile bmdifficulty

Eebit:
I'm probably missing something very silly on the remaining non-matches. 

I couldn't figure out what to do to match `sub_8037E4C` and `sub_8037E64`, which appear to be the save/load functions for the "record" / permanent Tower/Ruins data. CE link for `sub_8037E4C` which was as close as I could figure out: https://cexplore.karathan.at/z/-hA0TU

There's also a non-match / fake-match in `sub_8038F78` that got resolved with a raw ASM call. I think I got rule 5'd on this.


src/bmdifficulty.c
```diff
@@ -0,0 +1,1625 @@
...
+
+// SaveDungeonRecords?
+void sub_8037E4C(struct Dungeon* savedDungeon) {
+
+    struct Dungeon* dungeon = &gDungeonState.dungeon[0];
+
+    savedDungeon[0] = dungeon[0];
+    savedDungeon[1] = dungeon[1];
+
+    return;
```
jiangzhengwenjz, line 272: `memcpy(savedDungeon, gDungeonState.dungeon, 2 * sizeof(struct Dungeon));`

Eebit: Thanks! :)

PR #161, opened 2022-08-02
Dump a lot of proc scripts - trying to get closer to shiftable

NoahNelson:
Going to make this more automated and find/dump a bunch more.


ProjectRevoTPP:
Is this meant to be a draft PR or ready to go? I can mark it as one.

NoahNelson:
It matches, so could be merged, or you could wait until I dump the rest of the scripts. Doesn’t really matter 🤷‍♀️

NoahNelson:
better mark it draft actually - more coming soon :)

NoahNelson:
Ok, this is ready now. should be most proc scripts that include pointers, I may have missed a few.

PR #245, opened 2022-12-22
(statbar) done

MokhaLeee:
Functions related to draw stats-bar gfx in stat-screen.

non-matching function: sub_8086A40
https://decomp.me/scratch/rcQ8c


jiangzhengwenjz:
https://decomp.me/scratch/3gwZ3

MokhaLeee:
Force updated and removed that non-matching function.

PR #347, opened 2023-05-09
mapanim-lvup & lvupfx: decompiled

MokhaLeee:


MokhaLeee:
one non-matching function:

https://decomp.me/scratch/J1oVn

PR #387, opened 2023-06-14
match UpdateBanimFrame and improve sub_8059E18

jiangzhengwenjz:
sub_8059E18 only fakematches with -O1 but I guess it's simply something that I failed to figure out.


jiangzhengwenjz:
The force-push is only for adding back one missing `const` qualifier. This PR is ready for review.

PR #447, opened 2023-08-23
Decomp Event37 (GiveItem)

CT075:
It matches, at least, but I think we might want to take a look at/discuss changing some of these names before merging, because `Event37_GiveItem` is not a realistic, human-chosen name.


CT075:
eh, unless someone has some meaningful review, may as well just merge it

src/eventcmd_37.c
```diff
@@ -0,0 +1,45 @@
...
+#include "popup.h"
+
+// TODO: Give this a more human name (EventCmd_GiveItem?)
+int Event37_GiveItem(struct EventEngineProc *proc) {
+    u8 currentEventLoBits;
+    struct Unit *target;
+    u8 currentEventLoBits_;
+    s32 gold;
+
+    currentEventLoBits = 0xF & *proc->pEventCurrent;
```
Eebit, line 17: @MokhaLeee had added a macro for getting the event subcommand in [`event.h`](https://github.com/FireEmblemUniverse/fireemblem8u/blob/e925c2210648b5383149d13b839cc8eaf2970a96/include/event.h#L122), maybe we could use it here?

```suggestion
    currentEventLoBits = EVT_SUB_CMD(proc->pEventCurrent);
```

CT075: Good idea, been changed. I also took the liberty of using a better name than `currentEventLoBits`.

PR #459, opened 2023-09-07
Revisit player phase and prep phase

Eebit:


src/playerphase.c
```diff
@@ -147,10 +150,14 @@ PROC_LABEL(3),
     PROC_WHILE(DoesBMXFADEExist),
 
     PROC_END,
-
 };
 
-struct ProcCmd CONST_DATA gProcScr_0859ACE8[] = {
+// clang-format on
+
```
RevoSucks, line 156: What was the on off for?

Eebit: I was wrapping each proc script individually with the toggles in case we wanted to move them around. But looking at it now it looks silly, so I removed the overuse of toggles in 5b43da4

PR #460, opened 2023-09-11
Decompile muctrl

Eebit:
One stubborn non-matching function in `MoveUnit_`: https://decomp.me/scratch/Pvrgy

It would be great to have other eyes on documenting some of this. As far as I know, `REDA` is a historical name (short for "Reinforcement Data" or "Relocation Data") so I left it as-is. If someone has a better suggestion for naming it, I'm all for it.


jiangzhengwenjz:
https://decomp.me/scratch/2oQbw

Eebit:
Thanks a lot, Jiang :)

PR #467, opened 2023-09-28
Decompile difficulty select menu

Eebit:
Also matched a few more `savemenu` functions.

In order to match `sub_80ABC14`, I had to extract the "config" / option bits part of the `PlaySt` struct into a separate struct.

There are 3 nonmatches as part of this PR:
* `sub_80ABE3C`: https://decomp.me/scratch/pmTo7
* `sub_80ABF74`: https://decomp.me/scratch/SDtUT
* `sub_80AC4F8`: https://decomp.me/scratch/GaYjI
  * Seems to be responsible for the "glowing" effect of the difficulty selection that is being hovered over. I think it works with a 3D array of palettes, but I'm not 100% sure.


jiangzhengwenjz:
https://decomp.me/scratch/v1uik
https://decomp.me/scratch/HauTc
https://decomp.me/scratch/mKrgi

I really recommend you to use the permuter, since all the 3 functions are very easy to solve with it.

MokhaLeee:
what is the so called “permuter"?

Eebit:
Thanks a lot, @jiangzhengwenjz! :)

PR #470, opened 2023-10-02
Decompile CG text handler

Eebit:
Initially, I was calling this `serif` based on the FEBuilder label of the proc at `FE8U:0x08A01DBC`. It seems like this is the logic for handling CG text.

I tried to document the flags based on what I could reason about from the code, but I think it will take some more trial and error to figure out the actual purpose of a few of them since this is used in so few places in vanilla FE8.


laqieer:
I think `serif = セリフ = せりふ = 台詞 = CG text`.

PR #471, opened 2023-10-08
Gameintrofx

MokhaLeee:
https://decomp.me/scratch/lfh8N


jiangzhengwenjz:
I saw that this was matched, but you didn't match the check at the beginning properly. It's an optimization for checking negative value. https://decomp.me/scratch/PCI53

PR #514, opened 2023-11-26
rename gBmSt.unk3D --> gBmSt.um_mask and add enums

MokhaLeee:


MokhaLeee:
https://feuniverse.us/t/mokhas-hacks-sundries/16204/34

MokhaLeee:
In vanilla, if unit has traded with somebody, he can handle other action freely but can only wait if player want to control other units. This taken action is saved in gBmSt + 0x3D, and may casue wait during pre-action handler PlayerPhase_PrepareAction

PR #517, opened 2023-11-28
review event-battle, menu-panel, evtsub, unit-action

MokhaLeee:


src/cp_0803E2F4.c
```diff
@@ -257,11 +257,11 @@ s8 AiTryHealSelf(void) {
                 struct Vec2 position;
 
                 if (AiFindSafestReachableLocation(gActiveUnit, &position) == 1) {
-                    AiSetDecision(position.x, position.y, 6, 0, i, 0, 0);
+                    AiSetDecision(position.x, position.y, UNIT_ACTION_STEAL, 0, i, 0, 0);
```
Eebit, line 260: The `AiSetDecision` function should be using a different enum for the decision state: https://github.com/FireEmblemUniverse/fireemblem8u/blob/10673bf07ba423de6716b0ecf28c57fa910b236c/include/cp_common.h#L62-L78

MokhaLeee: Thanks! now updated

PR #548, opened 2024-01-09
dump event ch2~ch10a

MokhaLeee:
add auto dump python script (WIP), events need to be further interperted as is auto-dumped for now


MokhaLeee:
decide to move this PR to draft as I consider to dump more data

PR #569, opened 2024-02-07
Port unitlistscreen from FE6

Eebit:
Big thanks to @StanHash for all of their hard work and for tackling The Big Func (`SortUnitList`).

Two non-matching scratches, which are the same ones that don't match in FE6. They seem to be the functions for handling the graphics while swapping pages in and out.
* `sub_8091D54`: https://decomp.me/scratch/lXNoY
* `sub_8091F10`: https://decomp.me/scratch/vSnVv


jiangzhengwenjz:
The first one is easy. https://decomp.me/scratch/lXNoY
I can check the other one tomorrow but it looks terrible...

jiangzhengwenjz:
I can't match the second one but I believe most problems are resolved. https://decomp.me/scratch/3RUUz
In the 3rd loop, we need to prevent the constant 8 from being unrolled. But how?

PR #595, opened 2024-03-21
[Tweak] mint/name gpARM_HuffmanTextDecomp

CT075:
This serves twin purposes of bringing another CLib name to decomp and also removing an ugly `gUnknown_` name.


MokhaLeee:
I recommand to pick up StanH's solution in fe6:

https://github.com/StanHash/fe6/blob/main/src/ramfunc.c#L26

PR #607, opened 2024-04-30
rename statbar functions and params

CT075:
title

thanks to zane for double-checking a bunch of these parameter names


Eebit (approved):
Looks great - much appreciated!

PR #608, opened 2024-04-30
decomp all event related data

MokhaLeee:
1. dump all eventinfos
2. dump event shoplist/trapdata/chest-items
3. interpret ending event in game-ctrl
4. split messed eventscript


src/bmtrap.c
```diff
@@ -254,38 +254,36 @@ void LoadTrapData(const struct TrapData * data)
...
                     break;
 
                 case TRAP_MINE:
                     AddTrap(data->xPos, data->yPos, data->type, 0);
                     break;
 
                 case TRAP_LIGHTARROW:
-                    AddArrowTrap(data->xPos, data->unk_04, data->unk_05);
-                    // break; // BUG?
+                    AddArrowTrap(data->xPos, data->turn_counter, data->turn);
```
Eebit, line 282: Could you update this to use `#if BUGFIX` here? It seems that this was an omission of the FE8 devs when adding a new switch case to this function.

This was encountered as a bug in the Discord: https://discord.com/channels/144670830150811649/470029781795078175/1219122326122135643

Eebit (changes requested):

src/bmtrap.c
MokhaLeee: updated

PR #624, opened 2024-06-09
Modern GCC and GDB

dnasdw:
Support compilation with modern GCC and enable source-level debugging using GDB (e.g., in VSCode with mGBA).


Eebit (commented):
These are some good changes and much appreciated to have some of these things cleaned up! I've left my preliminary thoughts about some of the fixes in the actual decompiled code.

For the changes related to the build system, I think it would be best to have them reviewed by @laqieer, @CT075, @StanHash, or @MokhaLeee (who each have a better grasp on the various aspects of it than me).

Makefile
```diff
@@ -11,20 +24,37 @@ UNAME := $(shell uname)
...
+  CPP := $(PREFIX)cpp$(EXE)
+endif
 AS := $(PREFIX)as$(EXE)
 LD := $(PREFIX)ld$(EXE)
 OBJCOPY := $(PREFIX)objcopy$(EXE)
 STRIP := $(PREFIX)strip$(EXE)
 
-CC1     := tools/agbcc/bin/agbcc$(EXE)
-CC1_OLD := tools/agbcc/bin/old_agbcc$(EXE)
+# note: the makefile must be set up so MODERNCC is never called
```
CT075, line 47: Can you give more details on what this is accomplishing?

laqieer:
Thanks for your contribution. Could you please provide these to help us review this PR:

- [ ] resolve the conflicts with target branch first
- [ ] pipeline result to prove it passes and breaks nothing
- [ ] information on your testing environment (platform, OS, toolchain)
- [ ] result of `make` command to prove it builds with agbcc and matches the vanilla ROM
- [ ] result of command to build with modern gcc such as `make modern` and prove it works on hardware/emulator
- [ ] result of command to build with debug info such as `make modern DINFO=1` and proof of source-level debugging (also gdb version and emulator version)
- [ ] add related instructions to README

Makefile
dnasdw: Mainly refer to https://github.com/pret/pokeemerald

src/bmunit.c
```diff
@@ -989,7 +989,7 @@ void UnitChangeFaction(struct Unit* unit, int faction) {
         GetUnit(newUnit->rescue)->rescue = newUnit->index;
 }
 
-inline s8 CanUnitCrossTerrain(struct Unit* unit, int terrain) {
+inline s32 CanUnitCrossTerrain(struct Unit* unit, int terrain) {
```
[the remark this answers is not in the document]
dnasdw: This is likely another bug in the official code. Theoretically, there should only be a version of the function with a return type of s8. However, under the previous approach—where the function was not declared in the header file (.h) but individually declared in each source file (.c)—modern gcc would fail to link, unable to find the symbol for this function.

The current modification, while not ideal, ensures both the binary consistency after compilation with agbcc and resolves the linking issue with modern gcc.

Makefile
CT075: What parts of that repo is this intended to imitate? Should I believe that this is a good thing solely because emerald does it?

dnasdw: This change is mainly inspired by the Makefile section, which allows for the compilation of different versions of the code. Intermediate files and .o files are generated in separate directories to ensure that switching between different versions during compilation does not cause interference.

dnasdw:
image
image

I performed a rebase and then made corresponding additions in ldscript_modern.txt based on the changes in ldscript.txt today. I removed two inappropriate comments and added two new comments elsewhere to explain why the changes were made.

There are no conflicts now, and I manually compiled both versions without any issues.
The testing environment includes MSYS2, WSL1 (Ubuntu-22.04), and the operating system is Windows11, with the toolchain being devkitARM.
To compile the modern version of gcc, you can use either `make modern` or `make MODERN=1`.
There's no need for parameters like `DINFO=1` because the project's compilation parameters already include `-g`, so the compiled code is debuggable. However, when the `-O2` optimization is enabled, the game freezes during the map and story background description, and removing `-O2` solves the issue. Even pressing the START button to skip also leads to freezing, likely due to some animations.

As for README and similar files, it's best for your team to write them as I'm not particularly skilled in handling such textual content.

dnasdw:
> Thanks for your contribution. Could you please provide these to help us review this PR:
> 
> * [ ]  resolve the conflicts with target branch first
> * [ ]  pipeline result to prove it passes and breaks nothing
> * [ ]  information on your testing environment (platform, OS, toolchain)
> * [ ]  result of `make` command to prove it builds with agbcc and matches the vanilla ROM
> * [ ]  result of command to build with modern gcc such as `make modern` and prove it works on hardware/emulator
> * [ ]  result of command to build with debug info such as `make modern DINFO=1` and proof of source-level debugging (also gdb version and emulator version)
> * [ ]  add related instructions to README

Forgot to quote this, I mentioned the relevant content in my reply above.

PR #671, opened 2024-09-22
Small cleanups in Talk_OnIdle, GetStrTalkLen, BmBgfx_Loop, and StartGmapRm

minirop:
I skimmed through the code and notice a handful of cleanups.

`BmBgfx_Loop`: it can't be a switch statement, since you can't break out of 2 in C (except with goto) and return prevents the last 2 lines from being executed, so I removed the tentative switch statement.


RevoSucks:
lgtm

PR #672, opened 2024-09-23
Removed goto from some functions

minirop:
For SearchAvailableEvent, goto is still needed to match instead of break, but I rewrote the function so the flow is clearer.


Eebit (approved):
Great stuff! Thanks for this!

PR #674, opened 2024-09-26
Deciphering the goto pattern

minirop:
This isn't really meant to be merged (but could be, if you find the code cleaner for instance).

I rewrote those three functions to limit the number of goto's and for the goto's be close to their target, and I ended up with the same pattern for all 3, which basically boils down to:

```c
if (cond1)
{
    // some code...
    if (cond2) // or "} else {"
    {
        goto else_stmt;
    }
}
else
{
else_stmt:
    // some code
}
```

I have no clues how a compiler could generate that flow without a goto, so I'm asking here if someone might have an idea or two.


MokhaLeee:
I feel this modification is valuable. Can it match correctly now? If so, I think it is nice to merge it.

PR #675, opened 2024-09-29
Fix tools makefile

MokhaLeee:
In addition to the tools compiled with C, there are also some Python scripts that do not need to be compiled. They do not need make.


laqieer:
An alternative is to move `gfxtools` from `tools` to `scripts`.
- `tools`: tools writtern in C which needs compiler
- `scripts`: scripts written in Python which needs interpreter

MokhaLeee:
okay

Eebit (approved):

PR #676, opened 2024-09-29
sub_808679C matched, DisplayExtendedSysHand less ugly match

minirop:


src/sysutil.c
```diff
@@ -277,17 +277,10 @@ void DisplayExtendedSysHand(struct SysHandCursorProc * proc)
...
 #if !NONMATCHING
-    u32 clk;
-    u16 * src, * dst, * _dst;
-
-    clk = GetGameClock();
-    dst = gPaletteBuffer;
-    _dst = dst + (proc->pal_bank * 0x10  + 0x10E);
-    src = &PAL_BUF_COLOR(Pal_08A1D448, gPlaySt.config.windowColor, (clk / 4) % 0x10);
-    *_dst = *src;
+        ((gPlaySt.config.windowColor << 4) + ((GetGameClock() / 4) % 0x10))[Pal_08A1D448];
```
Eebit, line 282: I think it would be reasonable to remove the `if !NONMATCHING` here, since this style of accessing the array is used elsewhere in the codebase (especially in the world map logic)

src/events/unused-eventscript.h
```diff
@@ -17,7 +17,7 @@ CONST_DATA EventListScr EventScr_UnkCh3B_EndingScene[] = {
     ENDA
 };
 
-CONST_DATA u8 gAutoUdefJids[8] = {
+CONST_DATA u8 gAutoUdefJids[2][4] = {
```
Eebit, line 20: If you don't mind, could you also update the contents of the array to use the constants defined in `include/constants/classes.h`?

Eebit (approved):

Eebit (approved):

PR #683, opened 2025-02-04
Fix a mismatch on signedness of args on `DrawHpBar` (and friends)

Eebit:
As pointed out by @JesterWizard in the FEU Discord, there was a mismatch in the signedness of the arguments of the `DrawHpBar` (`FE8U:0x0808BC94`) function, and its associated helper functions. They should be using `u16 *` based on the type of `gUiTmScratchA`. Additionally, there was another mismatch on the `PlayerInterfaceProc` `unk_40`, which should also be unsigned. 

This PR makes the pointers `u16 *`, which is consistent with the FE6 decompilation: https://github.com/FireEmblemUniverse/fireemblem6j/blob/15073c1046053b7c1da6e5fc12237166c2c2fd7a/src/mapui.c#L193

Reference: https://discord.com/channels/144670830150811649/179027738454261760/1336121718527229952

### DIFF
--- a/src/player_interface.c
+++ b/src/player_interface.c
@@ -347,34 +347,34 @@ int GetCursorQuadrant() {
     }
 }
 
-void GetHpBarLeftTile(s16* buffer, s16 hp, int tileBase) {
-    if (hp > 5) {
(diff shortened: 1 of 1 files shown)


Eebit:
Superseded by #691

PR #689, opened 2025-02-14
Code review on `uichapterstatus`

Eebit:


include/fontgrp.h
```diff
@@ -228,11 +228,11 @@ void EndGreenText(void);
 // ??? GetSpecialCharChr(???);
 void PutSpecialChar(u16 * tm, int color, int id);
 // ??? PutNumberExt(???);
-void PutNumber(u16 *a, int b, int c);
-void PutNumberOrBlank(u16* a, int b, int c);
-void PutNumberTwoChr(u16 *a, int b, int c);
-void PutNumberSmall(u16* a, int b, int c);
-void PutNumberBonus(int a, u16 *b);
+void PutNumber(u16 * tm, int color, int number);
```
CT075, line 231: What is `tm` here? From context, maybe something like `canvas` or `buffer` might be more appropriate?

include/uichapterstatus.h
```diff
@@ -2,26 +2,26 @@
 #define GUARD_UICHAPTERSTATUS_H
 
 void StartChapterStatusHelpBox(ProcPtr);
-// ??? sub_808DCD0(???);
+// ??? GetFirstValidBlueUnit_Unused(???);
```
CT075, line 5: This doesn't need to happen in this PR, but can we start to expose these functions in these headers?

include/fontgrp.h
Eebit: `tm` is generally used to indicate `tile map` (in the sense referring to the buffer that gets copied to the hardware background) -- @StanHash has a definition for it in the FE6 decomp's [`GLOSSARY.md`](https://github.com/FireEmblemUniverse/fireemblem6j/blob/main/docs/GLOSSARY.md#tmtilemap)

include/uichapterstatus.h
Eebit: Good idea - added in 67ec833 :)

include/fontgrp.h
CT075: hmm, if that's the convention used elsewhere then i guess it's fine. I'd much rather just write out `tilemap`.

CT075 (approved):
happy to merge if nobody has any complaints in the next few hours

include/fontgrp.h
Eebit: I'll keep this in mind for future cleanup/refactoring!

MokhaLeee (approved):

PR #695, opened 2025-03-05
Add background enums and related macro calls

BlueAnthem37510:
As title says, replaces all background index mentions with enums instead. 
Also converted the rest of "Event_TextWithBG" and "EventScr_SetBackground"  calls into their [macros](https://github.com/FireEmblemUniverse/fireemblem8u/blob/master/include/EA_Standard_Library/Convo_Helpers.h)

Please let me know if there are any instances I missed :)


src/eventscr.c
```diff
@@ -1584,8 +1585,8 @@ void sub_800EC50(struct ConvoBackgroundFadeProc * proc)
             } // oh
 
         case 1:
-            if (proc->bgIndex == 0x37) // TODO: use an enum for convo backgrounds
-                proc->bgIndex = NextRN_N(0x35);
+            if (proc->bgIndex == BG_NONE) // TODO: use an enum for convo backgrounds
```
Eebit, line 1588: You can remove the `TODO` comment here since it has been addressed :)

include/constants/backgrounds.h
```diff
@@ -0,0 +1,63 @@
...
+    BG_RENAIS_CHAMBER          = 0x2E,
+    BG_WHITE_CHAMBER           = 0x2F,
+    BG_DESERT                  = 0x30,
+    BG_DARKLING_WOODS          = 0x31,
+    BG_VOLCANO                 = 0x32,
+    BG_BLACK_TEMPLE_OUTSIDE    = 0x33,
+    BG_BLACK_TEMPLE_INSIDE     = 0x34,
+    BG_BLANK                   = 0x35,
+
+    BG_NONE                    = 0x37, //Unsure?
```
Eebit, line 60: I might suggest something like `BG_RANDOM_SUPPORT_VIEWER` (or something less verbose -- names are hard), since it's used to pick a random background from the 0x35 real backgrounds when viewing a support

Eebit (approved):
Thanks a lot! The macro clean up in the event scripts is a huge help as well!

Just a couple minor points for your consideration added as inline comments :)

include/constants/backgrounds.h
Eebit: Having said that, maybe `BG_RANDOMIZED` or `BG_RANDOM` is sufficient to explain it -- it's used in vanilla for the support viewer but there's nothing to say you couldn't use it in a different context

MokhaLeee (approved):

src/eventscr.c
BlueAnthem37510: Knew I was going to miss one!

MokhaLeee (approved):

PR #697, opened 2025-03-08
Replace FIDs in text to names

BlueAnthem37510:
As title says. Set the case to match the rest of the definitions, but happy to adjust.


BlueAnthem37510:
Forgot to say, names are from [portrait_names_list.txt](https://github.com/FireEmblemUniverse/fireemblem8u/blob/master/scripts/portrait_name_list.txt)

Eebit (approved):
Once again, thank you very much for your efforts here! :)

I'd be happy to accept this PR as-is. 

One thing for consideration is that we have an incomplete enum in `UPPER_SNAKE_CASE` as [`include/constants/faces.h`](https://github.com/FireEmblemUniverse/fireemblem8u/blob/master/include/constants/faces.h) -- it currently only covers (some of) the special cases that are defined explicitly in the code, but the case is inconsistent with the values used in the text.

I see that commit a9e6ccc was made to make the case match the existing definitions in `textdefs`, so I'd want to get opinions from other contributors on whether it's worth making everything match the enum we have already. @MokhaLeee / @CT075 - do you have particular opinions?

BlueAnthem37510:
@Eebit It's no problem really, I'm happy to help! :)

Yes, I wasn't quite sure which case would be appropriate here. 😵‍💫 My personal reasoning was to make it match the rest of the text around it, but I agree they are more like enums here than the other text definitions.

MokhaLeee (approved):

MokhaLeee:
I think this PR is suitable for merging👍

PR #702, opened 2025-03-21
Some cleanup

minirop:
Among another round of typo cleaning, I managed to functionally match ParseBattleHitToBanimCmd.
regalloc is wrong on a handful of lines (see decomp.me link), and couldn't get rid of the ugly `r8++; r8--;` hack.


src/banim-battleparse.c
```diff
@@ -314,28 +314,26 @@ const u16 gUnknown_080DAEE6[EKR_DISTANCE_MAX] = {
 
 #if NONMATCHING
 
+// https://decomp.me/scratch/kcZgC
```
MokhaLeee, line 317: This function has been matched to 99.99% before. While I have no objection to improving the readability of this unmatched function, but it has only 99.89%. I recommend continuing jiang's work to maintain the higher percentage work.

https://decomp.me/scratch/meWtV

minirop: I started from it (I didn't find the decomp.me link, I copy/pasted the code from the repo), but it has some `asm` statements, if you remove them and this *syntax* `+ ({round_sp0C * 2 + 1;});`, it goes under 99%.
I'm still trying to fix the regalloc but to no avail ATM.

MokhaLeee: @jiangzhengwenjz @Eebit 
How about your opinion?

Eebit: I think the readability improvements here are good enough to merit a slight decrease in pure matching percentage. Especially since it gives people the ability to search the codebase for usages of the `BATTLE_HIT` and `ANIM_ROUND` enums, which has come up before in the FEU Discord

src/banim-battleparse.c
```diff
@@ -522,7 +502,7 @@ void ParseBattleHitToBanimCmd(void)
 
                     r9 = r9 + 1;
                     gEfxHpLut[r9 * 2] = new_hp;
-                    *anim_round_r5 = (u32)*anim_round_r5 | ({0xFFFF8000 + 0;});
+                    gAnimRoundData[round_sp0C * 2] |= 0x8000;
```
Eebit, line 505: Should this be [`ANIM_ROUND_DEVIL`](https://github.com/FireEmblemUniverse/fireemblem8u/blob/eda9ed07af7e52b77c68a194d3a0082f39a69b5a/include/ekrbattle.h#L287)?

minirop: Fixed.

Eebit (approved):

src/banim-battleparse.c
jiangzhengwenjz: > jiangzhengwenjz Eebit How about your opinion?

Sorry, I didn't see the message until now. I'm definitely not against it; my scratch was just a last-ditch attempt to match the function somehow. But it failed. ;_;

> I started from it (I didn't find the decomp.me link, I copy/pasted the code from the repo)
You can search for scratches by function label on decomp.me

PR #705, opened 2025-04-06
Add chapter enums

BlueAnthem37510:
Also some message id enums


MokhaLeee (approved):

src/savemenu.c
```diff
@@ -246,7 +247,7 @@ void SaveMenu_SetLcdChapterIdx(void)
 
         if ((gGMData.state.raw & 3) == 3)
         {
-            if (chapterId > 0x01 && chapterId != 0x38)
+            if (chapterId > 0x01 && chapterId != CHAPTER_CASTLE_FRELIA)
```
Eebit, line 250:
```suggestion
            if (chapterId > CHAPTER_L_1 && chapterId != CHAPTER_CASTLE_FRELIA)
```

src/savemenu.c
```diff
@@ -256,8 +257,8 @@ void SaveMenu_SetLcdChapterIdx(void)
             }
         }
         else if (gPlaySt.chapterIndex == 0x06 && CheckFlag(EVFLAG_136) != 0)
```
Eebit, line 259:
```suggestion
        else if (gPlaySt.chapterIndex == CHAPTER_L_5 && CheckFlag(EVFLAG_136) != 0)
```

src/worldmap_path.c
```diff
@@ -1865,9 +1865,9 @@ int GetPlayChapterId(u32 chapterId)
     int i;
 
     if (chapterId - 0x25 < 9)
```
Eebit, line 1867: It may be worthwhile for us to consider converting this into a macro for `CHAPTER_IS_TOWER` (and similarly below, `CHAPTER_IS_RUINS`), but not necessary for this PR

src/bmdebug.c
```diff
@@ -247,8 +249,8 @@ u8 EndMenuAndClear(struct MenuProc* menuProc, struct MenuItemProc* menuItemProc)
 }
 
 int CONST_DATA gTextIds_OnOff[] = {
-    0x0849, // TODO: msgid "ON"
-    0x084A, // TODO: msgid "OFF"
+    MSG_849, // TODO: msgid "ON"
```
Eebit, line 252: Not a strict requirement, but the `TODO` part can be removed now :) 

I have been leaving the message text inline since the numeric identifiers don't help at a glance, but if there is a better solution I'm all ears

Eebit (approved):

src/bmdebug.c
BlueAnthem37510: Done! :)

Eebit (approved):

PR #710, opened 2025-04-09
fix banim related wrong macros

MokhaLeee:


src/banim-ekrmain.c
```diff
@@ -69,11 +69,11 @@ const u8 BanimDefaultModeConfig[ANIM_ROUND_MAX * 4] = {
 };
 
 const u8 BattleTypeToAnimModeEndOfDodge[5] = {
```
Eebit, line 71: It seems the symbol name doesn't match what the enums are describing with this change -- does it make sense to rename the symbol as well?

PR #722, opened 2025-08-24
Dump opening subtitle graphics

BlueAnthem37510:


MokhaLeee:
The image obtained from this dump is confusing. Is there a way to replace it with an image that is easy to understand and edit, and build a script to extract the TSA from the image?

MokhaLeee:
@Eebit what's your opinion on this PR?

BlueAnthem37510:
Hi @MokhaLeee I've made some progress on an image to 4bpp + tsa convertor.

I'm away the rest of the week, so I thought I would just push what I have so far.

It doesn't include support for multiple palettes, and I'd like to rewrite it so it can be integrated into tsa_generator.py.

I'm having a little trouble with Op_Subtitle_05 as the baserom image without the map has extra empty space at the bottom which my script doesn't include, meaning the tile outputs don't quite match each other.

BlueAnthem37510:
Hello! Pardon the delay on getting back to this.

The convertor can now take any png (that's palatted correctly) and convert into the tiles + map, and I've converted most of the background images to use it.

However some of the games' graphics have a few quirks which makes it less than straight forward to binary match, which I have listed below.

**Arbitrary Starting Tile Index** 
    
   Some of the background images' tsas first index is a random number rather than starting at 0. For example bg_Grado_Chamber starts at 594 and bg_Burning_Castle starts at 29. I added functionally to specify the starting index in a json file to get around this. I might be missing some obvious pattern here though lol.
image
image


**Padding**

   A few have blank tiles at the start or the end of their tiles. For example Op_Subtitle_05 has 6 at the end, bg_Stream has 1 at the start, and bg_Renais_Chamber has 1 at the end.
    Same as before you can specify the number of padded tiles to add with a json file. Minus padding (e.g -1) puts the padding at the start instead of the end.
(bg_Stream)
image
image


**Skipped Indexes**

   Every other background I haven't converted because they skip indexes. I haven't checked every single instance where this happens but I don't think they are possible to binary match just by converting. 
    For example bg_Cell. Its tsa never uses index 16, however if you look at the nomap image data, tile 16 is not blank so when the nomap image and tsa are combined into the full image tile 16's data is just gone.
image
image


Happy to clarify anything here further, I think other than point 3 it's usable. :)

PR #732, opened 2025-12-31
UnknownMenu -> WeaponSelectMenu

ditto-salsa:
Just a name change for this menu as I noticed that it was still called "Unknown", when it is used for Weapon Selection upon attacking.


Eebit (approved):
Nice, thanks!!

